PR #3, opened 2016-03-21
xData rename to data

dreymonde:
As discussed in Slack.


loganwright:
I think a function here is less likely to have conflicts or provide alternatives to conformers. `data` is a pretty common word and I've seen it attributed to things like `[UInt8]`.

PR #11, opened 2016-04-09
Protocols for Queuing, Pooling, Scheduling, Handling, Jobs, and Workers

scottbyrns:


dreymonde:
My personal opinion is that it’s mostly related to specific styles of programming, and not to the Swift itself. If this will be included in SwiftX (or OpenSwift, whatever), I guess it should be in another repo, not in the `C7`.

PR #13, opened 2016-04-10
«.never as default» behavior and Data operators fix

dreymonde:
I personally don’t really like that with the newest version of `C7` we are forced to write `stream.send(data, timingOut: .never)` each time we don’t want to care about deadlines. We can’t have default values in protocols, but we have protocol extensions. This PR allows next: `stream.send(data)`
Protocols that were extended: `Sending`, `Receiving`, `AsyncSending` and `AsyncReceiving`.


paulofaria:
I think there are other APIs with deadlines. Think it would make sense to add defaults for them as well.

dreymonde:
Okey, I’ll update the PR then. Just wanted to make sure that idea itself is OK.

dreymonde:
Updated.

PR #29, opened 2016-05-08
Add Data initializers to prevent unnecessary copying

kevinup7:
Prevents performance issues when creating `Data` from an `ArraySlice`.

``` swift
let body: Data = getData()
let slice: ArraySlice<Byte> = parse(body: body)
let data: Data = Data(slice)
```

### DIFF
--- a/Sources/Data.swift
+++ b/Sources/Data.swift
@@ -22,10 +22,26 @@ extension Data {
     }
 }
 
+extension Data {
+    public init(_ slice: ArraySlice<Byte>) {
+        self.init(Array(slice))
+    }
+}
+
 extension Data: RangeReplaceableCollection, MutableCollection {
     public init() {
         self.init([])
     }
+
+    #if swift(>=3.0)
+        public init<S : Sequence where S.Iterator.Element == Byte>(_ elements: S) {
+            self.init(Array(elements))
+        }
+    #else
+        public init<S : SequenceType where S.Generator.Element == Byte>(_ elements: S) {
+            self.init(Array(elements))
+        }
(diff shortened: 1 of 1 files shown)


loganwright:
Can you elaborate on the performance issue this is preventing? Is there some sort of optimization in Swift when using `Array(slice)` over `Array(sequence)`?

kevinup7:
Sure, something like this takes around 32 seconds in Debug mode, and 0.65 seconds in Release mode.

``` swift
let bytes: [Byte] = Array(repeating: 1, count: 2_000_000)

let slice = bytes[0..<1_000_000]
let data = Data(slice)
```

Adding this brings it down to 0.38 seconds in Debug and Release mode, and fixes the issue completely for `ContiguousArray`.

``` swift
extension Data {
    public init<S : Sequence where S.Iterator.Element == Byte>(_ elements: S) {
        self.init(Array(elements))
    }
}
```

And adding this brings it down to 0.38 seconds in Debug mode, and 0.001 seconds in Release mode, which is what I would expect it to be.

``` swift
extension Data {
    public init(_ slice: ArraySlice<Byte>) {
        self.init(Array(slice))
    }
}
```

It seems like it shouldn't be necessary to have the specific initializer for `init(slice: ArraySlice<Byte>)`, but for some reason it appears to not quite getting optimized fully. It's possible I'm missing something.

paulofaria:
@kevinup7 awesome man!!! this is probably one of the causes of the 10x performance drop. @antonmes  check this out.

paulofaria:
LGTM!!!

PR #30, opened 2016-05-11
Updating for Swift 05-09

czechboy0:
Non-breaking changes, just fixes warnings.


paulofaria:
nooooooooo! so ugly the empty tuple ahaha. I prefer `(Void) -> Void`.

paulofaria:
also.. even if it's not breaking I think it should be bumped and tagged since the `.swift-version` file changed.

czechboy0:
Hmm, I don't think `(Void) -> Void` is a great way to write things, `Void` means no argument, which is the same thing as `()`. So `(Void)` is just unnecessary repetition. 😊 

Err, most packages won't need updating to 05-09, so I think we should keep this a patch version so that we don't have to update **every** other package that uses this.

antonmes:
@czechboy0 yeah, but it's easer to read. +1 for `(Void) -> Void`

antonmes:
also we could write 

``` swift
func flush(timingOut deadline: Double, completion: (() throws -> ()) -> ())
```

paulofaria:
ahhhhh my eyessss!!!

czechboy0:
Well, I'm not saying we should replace all `Void` with `()`, it is better in some contexts. But writing `(Void)` just feels wrong, even to the point I'd expect it to be a syntax error, because it means `(())`.

czechboy0:
So I suggest we only use `()` in these contexts, where it's a signature of a function and we're referring to the argument list.

paulofaria:
yeah but I don't care about what it is in "reality" I care about readability.. and pure symbols are not easy to read.

I don't feel super strong about using `(Void)` instead of `()` but I do prefer `(Void)`.

czechboy0:
Okay, there's a clear affinity to `(Void)`, so gimme a sec to change it (but let's not wait for Travis again please, I need to update a few more).

czechboy0:
Closing for #31.

PR #32, opened 2016-05-16
Adding JSON type

czechboy0:
Adding a JSON type, so that JSON libraries don't need to be large monolithic codebases with a lot of duplication (basic types, conversion between Swift and JSON types, manipulation convenience functions, etc). Instead, one library will be able to focus just on e.g. parsing, another on pretty printing, another on providing a BSON representation and so on. The goal is to make all of them compatible and composable.

### DIFF
--- a/Sources/JSON.swift
+++ b/Sources/JSON.swift
@@ -1,0 +1,15 @@
+
+public enum JSON {
+
+    public enum Number {
+        case integer(Int)
+        case double(Double)
+    }
+    
+    case object([String: JSON])
+    case array([JSON])
+    case number(JSON.Number)
+    case string(String)
+    case boolean(Bool)
+    case null
+}


paulofaria:
why not this?

``` swift
public enum JSON {
     case object([String: JSON])
     case array([JSON])
     case integer(Int)
     case double(Double)
     case string(String)
     case boolean(Bool)
     case null
 }
```

loganwright:
I don't think this object belongs in C7, but if it has to be, I would prefer Paulo's implementation.

I think this violates the principle:

> Only the code that is absolutely necessary for cross-project compatibility will be included.

czechboy0:
Because the JSON spec doesn't define two types for numbers, just one. I think we should respect that.

The reason why I want to keep Int and Double separate is so that if you receive a JSON object e.g. `{"myAge": 25}`, parse it into JSON and turn it back into data, you won't get `{"myAge": 25.00000}`.

czechboy0:
@LoganWright We discussed this with @paulofaria and agreed it probably should go into C7. Happy to hear why you think it's a bad idea.

bradhilton:
Yeah my vote is Paulo's implementation. Even though JSON doesn't formally distinguish between integers and floating point numbers, I think it's much more Swifty to break it up into `Int` and `Double`.

czechboy0:
@bradhilton My proposed type _does_ keep Int and Double separate, but also respects the JSON spec by having just one `number` type.

loganwright:
I don't think its necessary for C7 to function which I believe violates the original principles. Maybe you guys could outline the value it's adding?

The number thing of `{"number" : 25 }`, I think this should still parse as `25.0` and be used as a double. The addition of the duplicate number types adds unnecessary complexity to end user and I think isn't always accurate depending on what unknown 3rd party is doing the parsing. All we know according to the Json spec is that it's a number.

paulofaria:
@LoganWright you're talking about this actually, right?

``` swift
public enum JSON {
     case object([String: JSON])
     case array([JSON])
     case number(Double)
     case string(String)
     case boolean(Bool)
     case null
 }
```

loganwright:
@paulofaria both, if we're including Json, I think it should absolutely be your suggested structure.

My main concern is _why_ we're adding it in the first place. Json libraries aren't going to import C7 just for a Json struct, and I don't understand what the value add is.

paulofaria:
About the motivation:

> Only the code that is absolutely necessary for cross-project compatibility will be included.

Currently we have **no** cross-project compatibility for JSON. And I see value in having a common type for JSON. Don't you? @LoganWright

czechboy0:
@LoganWright This might be getting too philosophical, but mixing Int and Double just feels wrong, exactly because you might actually be creating **incorrect** results, due to the different representation of Int and Double in a computer. Int's are precise, while Double's only have a certain accuracy. I think that if you send me certain JSON object, all I do is parse it and echo it back to you, there should be **no** risk of the meaning to actually change - which it might if you convert an Int into Double, you might lose/cut off the actual value.

loganwright:
We can come back to the number thing although I feel pretty strongly that it shouldn't be `Int/Double` to answer this:

> And I see value in having a common type for JSON. Don't you? @LoganWright

No, I don't

90% of us are using a Json lib of some sort, Honza has one, GFX has one, Zewo has one, and all have their own struct. 

While I agree, it would be nice to have one, I don't think this is the way to solve it.

czechboy0:
@LoganWright Read the description of the PR for the reasoning behind adding this. Yes, cross-JSON-lib compatibility is **exactly** the reason why we want to add it. It stems from a concrete example, where my JSON lib Jay focuses on great parsing, while I'm not interested in writing all the boilerplate convenience functions around managing JSON, which can be done by someone else with a very good approach instead. We aim to make JSON libs composable.

Danappelxx:
I would personally vote for not including the `JSON` type, and instead move [StructuredData](https://github.com/Zewo/StructuredData) to OpenSwift. That's actually the whole goal of `StructuredData` - a standard that can bridge between different data types. I agree with @LoganWright, while we do all have our own versions of JSON, its more important (in my opinion) to build a common bridge between them rather than making them all the same.

loganwright:
This is supposed to be minimum viable, I think if we're going to include full json support w/ convenience functions etc., we're crossing the line into building another web framework.

czechboy0:
@LoganWright Sorry, you misunderstood. This file I just added is **all** I want to add into C7. But it will allow multiple authors to create compatible JSON libraries: one parser, one pretty printer, one for easy conversions etc.

paulofaria:
@LoganWright the proposal is only about the enum.. no more.. this would be the MVP imo.

bradhilton:
I think we should have both `JSON` and `StructuredData` in OpenSwift.

Sources/JSON.swift
```diff
@@ -0,0 +1,15 @@
+
```
Danappelxx, line 1: leading whitespace? this is the real issue here

loganwright:
Then I disagree even more, why are these here. Maybe in open swift, but not this repo. It's not necessary, the mandate is only what's necessary as minimum viable interface to facilitate communications between web framework's different components.

paulofaria:
@bradhilton I agree.. both with just the bare enums.

bradhilton:
While we're at it, `XML`? ;)

loganwright:
and csv

loganwright:
and yaml

czechboy0:
@LoganWright C7 is for Swift standards, if I'm not mistaken, not web frameworks (that's what S4 is for, isn't it?)

paulofaria:
Well OK could be other repo. But C7 comes from Core.. and I think JSON would fit here. Even because if you're gonna build a JSON parser you're gonna need `Data` which is here. So a JSON type already has to depend on `C7` if it wants to be Open Swift compatible. If we create another repo for this then the JSON implementation is gonna have to depend on C7 **and** this other repo.

loganwright:
You're not going to need Data unless you fully opt in, I think it's naive to expect Json libs to include the C7 library.

czechboy0:
Okay, can we bring it back to the issue at hand? I believe there should be a standard JSON struct, which JSON libraries could conform to and users could build their own utilities around, without having to rewrite everything if they change their parser. This serves that function, allows for composition of different JSON libs. 

Is anyone opposed to the **idea** of having a C7 standard for JSON? If not, we can get into the details. But let's go step by step.

bradhilton:
I think we need a standard. I don't know where it should go.

paulofaria:
Well, as I said.. if you're going to have a JSON library that is Open Swift compliant it would have to depend on `Data` anyway...

bradhilton:
Would it? It's just a definition, there won't be any other features.

goloveychuk:
Duplicating my idea from slack there
open-swift is for standarts. JSON is already standart. JSON library is implementations of standart. But instead we can write needed protocols. `serialize` `deserialize` e.g.

czechboy0:
@paulofaria Technically, it won't have to depend on Data, I think @LoganWright makes a point here. I would be fine with having a separate repo in open-swift for this JSON spec, even if it'll be super tiny, just this file. Would that make everyone happy?

paulofaria:
What I mean is... if I want to create a full fledged JSON library that is compatible with Open Swift. I'm gonna need a parser.. and that parser will have to depend on `C7` to be considered an Open Swift parser. So if my JSON lib want's to be Open Swift compliant it will already depend on `C7` so it makes all the sense in the world for `JSON` to be here.

loganwright:
why does it have to import C7 to be compatible? GFX Json parser works fine w/ open swift and doesn't touch C7

bradhilton:
`J9` just for `JSON` :P

loganwright:
w/ separate repo, as `number(Double)` 👍

czechboy0:
@LoganWright Haha cool, with the number I disagree, but let's not dilute this discussion with that (we'll have that elsewhere)

paulofaria:
Well.. sure it can work. But that doesn't mean it's compatible in the sense that it talks directly with libs that are Open Swift compatible.. using the `Data` representation of binary data. If I want a JSON lib that is Open Swift compliant.. it has to depend on Open Swift somehow.. and the parser would depend on C7 for Data.

czechboy0:
@paulofaria Yes, down the stack, the app itself will most likely import both, but up the dependency tree, you might have JSON-handling-only libs that will only touch J9, never C7. See what I mean?

paulofaria:
But ok.. you can have a JSON lib that doesn't have a parser.. then you wouldn't need to depend on C7.

loganwright:
Why is a Json lib open swift compliant at all? (pre J9 or whatever)

czechboy0:
@LoganWright What do you mean? Which library?

Danappelxx:
@LoganWright so that consumer libraries can feed it C7's `Data`.

paulofaria:
Yeah, but I think most of the times you'll want a parser.. haha. So for most cases we would have a JSON lib depending on C7 **and** this other module. I think only depending on the JSON enum would be the exception.

loganwright:
gfx, the original Json serializer, not sure why more came after it in the first place,

You can easily extend to incorporate c7, that's what we do, it's like 20 characters

paulofaria:
Because we want cross module compatibility.. that's why Open Swift was created.

loganwright:
```
    public init(_ value: Data) throws {
        self = try Json.deserialize(value.bytes)
    }

    public var data: Data {
        let bytes = serialize().utf8
        return Data(bytes)
    }
```

czechboy0:
Hmm... Okay. So everyone cool with open-swift having a new repo for the JSON spec? (I quite like the name `J9` now, any other suggestions?)

paulofaria:
If we're going to have another repo.. (which I really don't think it's necessary) I don't think it should be J9 only for JSON. It should be a repo for base types like JSON.

czechboy0:
Sure, so also for XML etc?

paulofaria:
A repo for higher level data representation. JSON, XML, MediaPack, StructuredData, etc..

czechboy0:
Formats... F3? 😆

Danappelxx:
I think we should leave the `9` number for when we have a module starting with `K`. `K9` 😛

loganwright:
Why would I import all of those for a Json library?

czechboy0:
@LoganWright To conform to a standard?

paulofaria:
`T9` maybe? @LoganWright cause importing a whole module for just an enum doesn't make sense.

paulofaria:
So we would have a repo for each new higher level type?

loganwright:
adding code for the sake of adding code seems worse than importing a module for an enum

Danappelxx:
I agree with @LoganWright... Why should `JSON` import something that contains specs for `XML`, `CSV`, `YAML`, etc.

paulofaria:
@Danappelxx So we should have a repo for each new higher level type?

Danappelxx:
Can we add only `StructuredData` with protocols like `StructuredDataConvertible`? I don't see any downsides.

czechboy0:
Guys, I understand you want to keep things _pure_, but let's be practical here. I don't think creating a new repo for every new format makes sense. And this is a pretty good compromise IMO.

czechboy0:
@Danappelxx StructuredData isn't the same thing as JSON though. I want a strict JSON type, compiler-checked. This is the way to do it.

paulofaria:
@Danappelxx  Cause `StrucuteredData` representation is not compatible with `JSON` 100% of the times.

paulofaria:
Yeah I agree with @czechboy0 

> StructuredData isn't the same thing as JSON though. I want a strict JSON type, compiler-checked. This is the way to do it.

loganwright:
I want a Json type too, but with number as Double only. It's parsing `"key" : 1` isn't always an integer and `"key" : 1.0` isn't always a double. 

It should always be double to acknowledge the uncertainty.

paulofaria:
@LoganWright @Danappelxx Seems to me the biggest problem you're having.. is depend on other stuff when you only need one stuff. Right? But if you don't care about the other types.. just do nothing. Even if we add types for all common higher level types is gonna be like 100 LOC at most.

czechboy0:
@LoganWright HEY! Stop jumping around, let's settle one thing first before we start another. The int/double thing will have its own PR, I promise you'll have a chance to present your views.

loganwright:
So, if we do this, where do we stop? Just at the enum? what about convenience accessors that almost all Json libs have, subscripting, etc.

czechboy0:
@LoganWright Yes, just the enum. The convenience stuff is exactly an opportunity for a JSON-lib to solve.

paulofaria:
Yeah the idea is to have other frameworks add the computed properties, parsers, serializers, etc.

Danappelxx:
No, my issue isn't about having too much in `C7`. My issue is that I don't think `JSON` belongs in `OpenSwift` at all. My "dream" would be for `StructuredData` to be the only representation of data that your code works with, and _that_ sounds like a standard to me. JSON is already a defined standard and I don't see the benefit of putting in `C7`.

tanner0101:
I'm with Dan. I just want `StructuredData` in C7.

loganwright:
Can we move StructuredData to open swift?

dreymonde:
I also support @Danappelxx on this

paulofaria:
@Danappelxx `StructuredData` is good for when you have to represent more than one higher level data abstraction in on type. But when you **only** have to deal with JSON. You need a JSON type.

czechboy0:
@Danappelxx I understand and I think StructuredData is cool, but let's be realistic here: everyone uses JSON, and there are many JSON libs which are incompatible at the moment. And again, StructuredData and JSON are not the same thing, so I don't think you can make up for the lack of a JSON Swift standardization by having StructuredData.

dreymonde:
@paulofaria all these «JSON-only» doesn’t make any sense to me, for example, because I don’t want to deal with JSON at all, but with `StructuredData` instead.

czechboy0:
@dreymonde Nobody's forcing you. We're trying to **allow** for people who **want** to use JSON, to be able to use multiple JSON libraries that each solve a different part of the problem.

paulofaria:
We should have JSON and StructuredData at `C7` or if you want to separate.. put them in another repo for the higher level types. But `StructuredData` and `JSON` do **not** solve the same problem. They are not the same.

dreymonde:
If we create a standard for JSON, there will be a lot of JSON libraries which will be completely useless for developers who use `StructuredData`, which is kind of a better practice than to stick with JSON only.

czechboy0:
@dreymonde Once again, JSON and StructuredData aren't the same, don't solve the same problem. I'm not even sure why we're talking about StructuredData here, as the **only** thing this PR proposes is to standardize JSON libraries.

paulofaria:
@dreymonde if you don't want to deal with JSON just don't use it. You're not thinking about the bigger scope. You're thinking about web apps and content negotiation. That's not what Open Swift is about. Open Swift is about standards for Swift in general. Some applications **need** to have a strict JSON representation. And there's no reason we shouldn't have standard just because some don't want to use it. If you don't want to use it. Just don't use it.

paulofaria:
@dreymonde you know me.. you know I love `StructuredData` for content negotiation in web apps. and I will always suggest its use for this kind of application. But we can't disallow other usages in other scenarios because of that.

dreymonde:
@paulofaria 

> You're thinking about web apps and content negotiation

That’s true, yes. Now I get it. Okay, I have no objections then.

czechboy0:
Cool, so now that we've cover many aspects, what's the consensus? I still believe the best place to add this would be here, in C7. We've seen in the conversation above that even though many of use are working on web framework stuff, C7 is about Swift standards. And I believe that the community could benefit from one small standard for JSON representation, which libraries could conform to and thus work together nicely, just like S4 does for web frameworks.

This change is strictly about JSON - is anyone still against adding this file to C7, thus allowing (but not forcing) people to take advantage of a standard JSON type?

paulofaria:
I vote for this representation in C7:

``` swift
public enum JSON {
     case object([String: JSON])
     case array([JSON])
     case integer(Int)
     case double(Double)
     case string(String)
     case boolean(Bool)
     case null
 }
```

loganwright:
I've come around, I'm open to including it here.

Would like to revisit the number part.

loganwright:
I think I'd prefer @czechboy0 over that one @paulofaria

czechboy0:
Okay, let's discuss the int/double number issue. I've outlined above why I believe separating int/double is necessary to preserve correctness of transferred data. And to also stay in line with the JSON spec, I created a nested enum for int/double.

dreymonde:
I vote for adding JSON here. It belongs here nicely. Although if we’ll decide to include other formats and `StructuredData` (maybe) in OpenSwift, we should reconsider this.

dreymonde:
I like @czechboy0 approach with unified number case, it’s good for standard. Although it’s pretty weird to use it as it is, I’m sure other libraries will add more convenient ways.

czechboy0:
Consider this: there will be convenience functions to extract the types nicely, so the fact that there's a nested enum will be quickly made pretty much invisible of users of JSON libs, because those libraries will dictate how they treat numbers in detail (conversions between int and double etc). I believe we should keep the separation here, because libraries can still treat them as the same thing. If we merge them here, libraries won't be able to tell, what the data was originally parsed as.

tanner0101:
I vote this one:

``` swift
public enum JSON {
     case object([String: JSON])
     case array([JSON])
     case number(Double)
     case string(String)
     case boolean(Bool)
     case null
 }
```

```
A JSON value MUST be an 
- object, 
- array, 
- number, or 
- string, or one of the following three literal names:
- false 
- null 
- true
```

https://tools.ietf.org/html/rfc7159

paulofaria:
I agree with @dreymonde, but as we have versioned API I think it should stay here for now. Open Swift is far from 1.0 (I suppose we all agree it will only happen when Swift 3 gets released) so we can postpone the decision of a new repo.

czechboy0:
@tannernelson Yes, but when you look into the RFC how it defines a number, there's no reason to believe it's a floating point number, which this would force.

Sources/JSON.swift
```diff
@@ -0,0 +1,15 @@
...
+public enum JSON {
+
+    public enum Number {
+        case integer(Int)
+        case double(Double)
+    }
+    
+    case object([String: JSON])
+    case array([JSON])
+    case number(JSON.Number)
```
dreymonde, line 11: Actually, I think compiler will be fine with just `case number(Number)`

paulofaria:
> Consider this: there will be convenience functions to extract the types nicely, so the fact that there's a nested enum will be quickly made pretty much invisible of users of JSON libs, because those libraries will dictate how they treat numbers in detail (conversions between int and double etc). I believe we should keep the separation here, because libraries can still treat them as the same thing. If we merge them here, libraries won't be able to tell, what the data was originally parsed as.

I agree with this, but having the nested enum just to keep "adherence" to the standard is kinda convoluted.

tanner0101:
```
 This specification allows implementations to set limits on the range
   and precision of numbers accepted.  Since software that implements
   IEEE 754-2008 binary64 (double precision) numbers [IEEE754] is
   generally available and widely used, good interoperability can be
   achieved by implementations that expect no more precision or range
   than these provide.
```

> Since software that implements IEEE 754-2008 binary64 (**double** precision) numbers [IEEE754] is generally available and widely used

Sources/JSON.swift
czechboy0, line 11: It will, this is for better readability. But I can remove it.

paulofaria:
I think the storage is one thing, but we need to research more about the real-life problems that could arise from sending an Int representation and receiving a Double representation.

czechboy0:
I just don't like losing information. And I see no reason to discard information if we can nicely keep the information with the type system. Again, you'll rarely work with this directly, it'll all be solved in the JSON libs. So look at it from a purely functional view, not from "I'd prefer to do less work" (because you won't work with it).

tanner0101:
> I just don't like losing information

What kind of information are you losing? The JSON spec clearly indicates that serializers should not expect better than double precision implementations. Open Swift should adhere to the spec, not what we think might be better or more advanced. Too much accuracy is a bad thing in some cases, such as measuring latitude and longitudes.

paulofaria:
@tannernelson the problem is if you have an API that parses to JSON and then serializers without touching it you would send `{"foo":1}` and receive  `{"foo":1.0}`. that could cause problems. I'm just saying we should investigate this.

czechboy0:
@tannernelson You lose information, because the sender might send two payloads, `{"foo":1}` and `{"foo":1.0}`, a JSON parser will parse it, and if it's a good one, it'll give you the first one as `[String: Int]` and the second as `[String: Double]`. 

The presence or absence of a decimal point dictates whether it's an int or a floating point number, so if you get "1" and "1.0" and your parser represents them as the same thing, you're losing information (because the sender saw it as two completely different types).

czechboy0:
Moreover, we're creating a top level, simple standard, which I believe should allow better resolution, if practical. Keeping ints and doubles separate doesn't hurt us and will benefit people who want to keep them separate. For people who want to treat them as the same thing, they'll use a library which provides such convenience, e.g. `.getNumber() -> Double`, which will convert the Int to Double if necessary. But I don't see a reason why we should prevent the first group of people from getting what they want, when we have a solution which can please both groups.

tanner0101:
I think what you're inventing is something we should call SJSON (Swift-JSON). But JavaScript Object Notation was defined in RFC 7159 and created for a language that can't even differentiate floats from ints. We're adding complexity to it that simply isn't required or expected by the specification.

czechboy0:
You are correct, it's not required by the JSON spec, but it's allowed (according to the spec, libraries are in charge of how numbers are represented), so here we chose to represent them in a way that preserves the information of how the sender intended the number to be represented (int/double). This is a feature, not a bug :) What do you see as a problem? What scenario are you thinking about where this will be worse?

tanner0101:
If it's not explicitly mentioned in the spec that integers and floating point numbers should be maintained between parsing and serializing, then we're portraying false levels of accuracy to the users of the enum.

Take the example where someone makes a User with an identifier of 1 and it's an integer. They use any arbitrary RFC 7159 JSON spec serializer to send it from their API.

Now they are using an Open Swift project and they want to receive this integer. Well, naturally they would check for `case .integer`. But if the serializer they used doesn't care about the difference between an integer and a double, because it's not required to, then their check for an integer will never pass and thus they have a bug. 

I've seen this a million times in PHP where serializers will turn ints to doubles or even to strings sometimes. JSON is really not as precise as you're making it out to be and we shouldn't be promising something to the user of this enum that we can't back up.

Danappelxx:
Just wanted to add that `{"foo":1.0}` is not natural JSON - type in `JSON.stringify({"foo":1.0})` into a javascript console and you will get the output of `"{\"foo\":1}"`.

czechboy0:
You're just arguing against using a strongly-typed representation of JSON in general, all this can be applied to the distinction between bools/strings/numbers, so I strongly disagree. I think we should keep as much information as we can from the received data, and let a library take care of how it treats it.

In your example, the user would use a library that has a `number() -> Double` convenience call, which does the conversion to Double. 

I believe we should use the type system to our advantage. If someone wants to convert all types to String, they can. But I don't want to prevent people (like me), who want to preserve the information.

Think about this: you map a parsed JSON onto an object, one field is "gamesPlayed". The server sends me the number "1.2", I ask for Int and it fails - but that's a good thing! I'll go, debug, see the server is sending nonsense, and fix it. 
On the other hand, if I had a always Double-casting type, I'd just take the input and cut it into an Int, getting the number 1. So the server sent 1.2, but I'm using 1, waiting for a nasty bug to creep in.

See what I mean? We can both create examples of where the other's method fails, so I believe we should preserve as much information for the user of the library to make that call, instead of us trying to do so here. 

If there isn't a great reason to take away a capability, I don't think we should take it away.

bradhilton:
Since there is no `Number` type in Swift it makes sense to map it to the Swift equivalent which is `Int` _and_ `Double`. Representing it as `case number(Double)` is less accurate.

czechboy0:
@tannernelson I understand you wouldn't love this solution, but would you be okay with it making it in anyway? You see, what you want to do is possible with my solution (enables us both to do what we want), but your solution prevents me from doing what I want (can't discern number types anymore).

Does anyone else have any final feedback? If you're just joining, please read the full thread so that we don't cycle in already discussed points :)

bradhilton:
Just going to officially say that my vote is for @paulofaria's representation:

``` swift
public enum JSON {
     case object([String: JSON])
     case array([JSON])
     case integer(Int)
     case double(Double)
     case string(String)
     case boolean(Bool)
     case null
 }
```

I think it is the cleanest and most Swifty, while still accurately reflecting the JSON spec.

czechboy0:
@bradhilton: So you dislike this? https://github.com/open-swift/C7/pull/32/files

bradhilton:
I don't think we need an `enum` representation of `Number`

czechboy0:
@bradhilton Right, but it's a good compromise between respecting the JSON spec and also keeping the number types separate. I've used a setup like that for months, and it works really well, so unless there's a big issue with it, I think I'll keep the proposal to this. Flattening the structure and putting double and int as base JSON types feels like too big of a departure from the true JSON types, if it makes sense.

bradhilton:
I think it isn't necessary. I disagree with having the `Number` type.

czechboy0:
Read my earlier comments about nobody working with this type directly, meaning it's only for the benefit of the wrapper code (even today, most JSON parsers represent the parsed structure in a type-safe manner, but only present you with convenience functions to take a String out, or an Array etc). So this complexity is only for communicating the distinction of the parsed number, but won't be felt by users at all. Which I think is a good balance.

antonmes:
``` swift
public enum JSON {
    case object([String: JSON])
    case array([JSON])
    case number(Double)
    case string(String)
    case boolean(Bool)
    case null
}

extension JSON {
    var integer: Int? {
        if case .number(let value) = self {
            return Int(value)
        }
        return nil
    }
}
```

czechboy0:
Not the same thing, as if I send 1.0, it's a double but this function would actually claim it's an int. Just because something is convertible to int doesn't mean it was intended to be used as int. But just read the whole thread, I've explained my reasoning.

antonmes:
if you send 1.0 then use .number, if you're using .integer - you know what you're doing. 
How would you know someone sent a Double from client-side javascript while it was serialized as `1`.
i like having 

```
case integer(Int)
case double(Double)
```

but it's not "standard".

paulofaria:
I vote for my JSON representation! hahaha, does it count? 😆

antonmes:
@paulofaria 
how do you propose to handle this `[{ "type": 1 },  { "type": 3.14 }]` when i'm expecting `Double`. 

``` swift
for item in json {
  switch item {
  case .integer(let value): // do things
  case .double(let value): // do things
  default: throw Error()
  }
}
```

like this?

## 

@czechboy0 or like this? 🙂

``` swift
for item in json {
  switch item {
  case .number(.double(let value)): // do things
  case .number(.integer(let value)): // do things
  default: throw Error()
  }
}
```

## 

i would prefer 

``` swift
for item in json {
  guard case .number(let value) = item else {
    default: throw Error()
  }
  // do things 
}
```

antonmes:
@czechboy0 or like this? 🙂
```swift
for item in json {
  switch item {
  case .number(.double(let value)): //do things
  case .number(.integer(let value)): //do things
  default: break
  }
}
```

paulofaria:
@antonmes it depends on the higher level lib implementation.. the higher level lib can have `json.double()`only if one prefers and never expose the integer.

``` swift
extension JSON {
  func double() -> Double? {
    switch self {
    case integer(let i): return Double(i)
    case double(let d): return d
    default: return nil
  }
}
```

antonmes:
so what the point having this in standard if we'll use different api anyway?

paulofaria:
Cause the important part will be cross-project friendly. Why have Open Swift if our web frameworks will have different API? Haha, is the same question.

antonmes:
So why you're not happy with `Double` if you're not going to work with it?
If it's just for cross-project communication? Why doing standard not standard then?
**cons**:
having non-standard json
having redundant parser checks
having redundant switch statements

antonmes:
i would prefer 
```swift
for item in json {
  guard case .number(let value) = item else {
    default: throw Error()
  }
  // do things 
}
```

paulofaria:
The problem is that we can have bugs because of that. And don't say it's non-standard JSON, cause it is. The way the type is represented internally and how it is stored is only relevant in terms of its capability to effectively represent the raw data values. Meaning.. if you have a JSON like so:

``` json
{
  "foo": -1.5
}
```

And you represent `-1.5` internally with two `Int64` one for the significand and one for the base and a `Bool` for the negative sign. It's still going to be standard JSON. Cause it is capable of representing `1.5`. Now it wouldn't be standard JSON if it wasn't capable of representing that. The parser checks won't be redundant cause JSON allows integer, floating points and exponential notation. All of which need different treatment in the parser.

``` json
{
  "foo": 1
}
```

``` json
{
  "foo": -1.5
}
```

``` json
{
  "foo": 1E10
}
```

On the consumer side you should never have redundant switch statements. This is a type that is supposed to be extended in a higher level library. That library will provide you with convenience computed properties. Even properties that could force convert from one type to another.

The main point about keeping two types for the number is about preventing bugs, that I can easily spot if we cast everything to double. We want things to adhere to the RFC (which it does) and also be bug repellent (which it does).

loganwright:
It supports those different types, yes, but they're not to be considered distinct according to Json spec. If Json spec said include `.` to indicate floating point, otherwise it's an Int, that's a different story. 

Here we'd be actively imposing our desires onto a spec that doesn't adhere to them. Json includes `number` which is basically a floating type because that's how Javascript handles it.

If we're writing all the json parsers on both side and want a custom type that represents Integer that way, it's fine for a personal lib. But for global standards, we can't control every json serialization on the web and should stick to standard spec.

For all the specificity, integer, etc. I'd rather stick w/ zewo/StructuredData or qutheory/Node and use a specific serialization there that allows better accuracy w/o json.

antonmes:
integer + double = workaround in the core + nonspec parser + workaround for user-friendly api
number = < spec > + workaround for user-friendly api

paulofaria:
Ok! Taking in consideration what you guys are saying about keeping the `number` case and specifically what @czechboy0 and @bradhilton said:

> Since there is no Number type in Swift ...
> @bradhilton
> 
> Because the JSON spec doesn't define two types for numbers, just one. I think we should respect that.
> @czechboy0

I really think we should create a `Number` type for `C7`. I think this makes everyone happy. Another plus is that the `Number` type can be useful for a plethora of applications. So my proposal is:

`Number.swift`

``` swift
public enum Number {
    case byte(Byte)
    case integer(Int)
    case double(Double)
    ...
}
```

`JSON.swift`

``` swift
public enum JSON {
    case object([String: JSON])
    case array([JSON])
    case number(Number)
    case string(String)
    case boolean(Bool)
    case null
}
```

And `Number` here should be self-sufficient (no need for external libs extensions), so we would add computed properties converting from one type to the other and what not. With this we can keep the `number` specified by the RFC and still be bug free.

loganwright:
I really don't think we should be defining number, particularly to include byte, that's not really Json spec. Json spec says double.

paulofaria:
@LoganWright this is definitely not for `JSON`. This is a `Number` type.. a type that represents a number. As I said this can be used for a whole lot of other things. I went to look for implementations in other languages and this is what I found for Java. 

https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/JsonPrimitive.java#L55

They use `Number` 

https://docs.oracle.com/javase/8/docs/api/java/lang/Number.html

This is definitely the best type to represent a number.. can we agree with that? And if `JSON` states a `number` primitive that is.. a number.. what better type than `Number`? `JSON` RFC **never** stated it has to be a double...

loganwright:
I disagree that it's the best type to represent a number, I don't think it's a necessary addition and only adds confusion to the json type. We have to remember that this is Javascript, not Swift.

RFC doesn't need to because javascript doesn't distinguish, it only has number type which is a double.

paulofaria:
Yeah javascript number type might be a double (I don't know javascript, I'll have to believe you there). But no, this is not javascript. This is Swift. Just as Java uses their `Number` type I think it makes all the sense in the world for a number to be `Number` in Swift. All languages have a number abstraction. If Swift doesn't have one and we need it (I'm telling you, we need it. We can't get away with just casting to `Double` in every scenario and not even thinking about the consequences), we should create it. And again `Number` is not for `JSON`. It would be used by `JSON` but it wasn't made for it. It's just a number abstraction. I wish we could get away with just `Double` but unfortunately we can't.

Swift is great because of it's type-safety and we should harness that, not seclude it. Keeping type information is a great thing and I can't understand why you're being so against it.

loganwright:
We're using it in Swift, but it _is_ javascript. Javascript object notation. Unless we can control every parser on every endpoint, then we should follow suit to the type we're importing. I don't think it's necessary to bring a new number type into it, but maybe that's the answer.

Let me ponder on the Number type and see if others way in there.

antonmes:
@paulofaria because in this case we are not keeping it, we making it up, because we can't distinguish `1` and `1.5` which came from client-side

paulofaria:
@antonmes I'm not sure I understand. Can you give me example?

antonmes:
so developer should decide, which type he expecting, and throw an error otherwise.
example: https://github.com/open-swift/C7/pull/32#issuecomment-219586570

paulofaria:
Sorry I still don't understand.. so you parsed `[{ "type": 1 }, { "type": 3.14 }]` the representation will be: `.array([.number(1), .number(3.14)])` what about it?

antonmes:
in another words: we can't keep the type because json doesn't.

Danappelxx:
In JSON, the string `{"number":1.0}` _does not exist_. If you serialize `["number":1.0]` it will pop out as `{"number":1}`. Therefore, if you then parse that same JSON and assume that `number` is an `Int`, you are making up type information. Hence, it would be more accurate to parse it as a `Double` at all times, rather than assuming that it is an `Int` if it doesn't have a decimal point.

paulofaria:
JSON doesn't keep the type? What do you mean.. the type is `Number`. That is the type.

paulofaria:
> In JSON, the string {"number":1.0} does not exist.

I can't understand this sentence at all haha

paulofaria:
> If you serialize ["number":1.0] it will pop out as {"number":1}

Where did you get this from? Isn't this implementation dependent?

Danappelxx:
@paulofaria https://github.com/open-swift/C7/pull/32#issuecomment-219566099

antonmes:
> Where did you get this from? Isn't this implementation dependent?

@paulofaria open devtools in your browser and paste it into the console `JSON.stringify({num: 1.0})`

paulofaria:
Well! I'm sold by the fact that `Java` uses `Number` to represent well... numbers.. in their JSON library (specifically Google). And as I said before.. just because javascript represents number as double it doesn't mean we **have** to as well in Swift. I really can't understand why you'd prefer having less type information using Swift. I'm telling you this could be a source of bugs.

loganwright:
I wouldn't prefer less type information, I'd prefer honest type information.

> I'm telling you this could be a source of bugs.

Adding the number type is a guaranteed source of bugs. As @Danappelxx and @antonmes showed in dev tools, 1.0 => 1 so would parse as an Integer instead of its true type a Double. This isn't more type information, it's _inaccurate_ type information. 

By using Double, as implied by the spec, we're at least honest about that uncertainty.

antonmes:
I would not refer to java api  😂 but.. they use it because java has it. Swift doesn't. 
We don't have implicit casts as well, and this is more about making up the type, not saving it.
A agree that types matter, and if you expecting integer - you should check the value and cast it or throw an error, but it should be done by developer, not by a guessing parser.

paulofaria:
The thing is `Double` is **not** implied by the spec. It is the dev tools implementation, simply because javascript **can't** do different. It doesn't mean we have to do it like javascript. Again, this is Swift. We wouldn't be misleading in any way. If I parse `{"foo":1}` and serialize it I expect  `{"foo":1}` not `{"foo":1.0}`. Just because javascript isn't capable of that doesn't mean that we should do it as well. I really really really don't know why you want to do the javascript way given what we have is Swift.

Danappelxx:
The spec is based on Javascript, is it not? Number in the context of the spec can be interpreted as a Javascript number, which is just a floating point (Double).

paulofaria:
Hahaha nope.. stop it.. you're pushing it. And I really can't understand why. The spec only talks about the precision.

antonmes:
as i said before, we should check the type anyway, the difference is that done by parser and Number type, ooor it can be done through convenience propeties (optional) which way is more swifty? 
https://github.com/open-swift/C7/pull/32#issuecomment-219579249

Danappelxx:
Not convinced? Apple does it the same way. (excuse my Swift 2.2)

``` swift
import Foundation
let dict = ["hello":1.0]
let data = try NSJSONSerialization.dataWithJSONObject(dict, options: .PrettyPrinted)
let json = String(data: data, encoding: NSUTF8StringEncoding)
print(json) // -> Optional("{\n  \"hello\" : 1\n}")
```

antonmes:
And again, we're not losing concrete number type because json doesn't provide  it. 
Conversely, we're splitting it in a different types and have to clue them later. https://github.com/open-swift/C7/pull/32#issuecomment-219586570

bradhilton:
@Danappelxx I'm not sure that's the best example, since Apple also wraps BOOL in it's `NSNumber` type. I'm sure if they reimplemented it now it would be different.

antonmes:
Also with the Number type we would definitely have [this](https://github.com/antonmes/MessagePack/blob/master/Sources/MessagePackType.swift#L70)  😂

loganwright:
As much as we'd like it not to, JSON uses Doubles. Anything else is dishonest and error prone. Two extremely ubiquitous serializers, Chrome, and Apple both send 1.0 as 1. Although it comes as 1 w/o a `.`, it is a Double and should be treated as such.

paulofaria:
@antonmes the issue is the parsing/serializing part.. 

> If I parse {"foo":1} and serialize it I expect {"foo":1}

But ok this thread has become too long. I think it's by far the biggest thread hahah. My point is `Number` is a more powerful type abstraction which is something that Swift can provide which allows us to keep the parsing/serializing stuff above. If we don't use `Number` and use `Double` we'll be just like javascript. And I can't understand why you prefer that since the `Number` approach is still RFC compliant and more powerful. Both approaches would need the same computed properties so in the end for force-casting to a given type you want to receive would need the same API. The difference is that with `Double` you lose the format information.. so it would be less powerful. I prefer `Number`, but I'm OK with `Double` if that's the majority. Really hope we don't have bugs. But still.. why you prefer the javascript way will still forever be a mistery... hahahahahaha 😈

loganwright:
Regardless of Number being useful, JSON is _not_ a `Number`, it's a `Double`. Perhaps Number has value, but should be discussed in another thread.

> Really hope we don't have bugs

Honesty about uncertainty is less error prone and will have fewer bugs than false certainty.

> why you prefer the javascript way will still forever be a mistery... hahahahahaha 😈

I don't prefer it, I accept it.

antonmes:
@paulofaria why javascript way? i don't want parser to decide for me is it integer or double. i think it's more error prone than having `the original value` and optional properties. without redundant `switch`.

paulofaria:
> JSON is not a `Number`, it's a `Double`.

In Javascript yeah. But Java, for example, no. See that's what I don't understand haha. Why you're using javascript as the definition of JSON. Just because javascript represents JSON numbers as `Double` it doesn't mean all languages should. JSON may have started with javascript but it's not javascript. The RFC never says it should be represented as `Double` it only talks about the precision. But this is getting far too old.

antonmes:
@paulofaria because Java doesn't have extensions and optional properties and we do. 😄

paulofaria:
@antonmes it's all about this:

> If I parse {"foo":1} and serialize it I expect {"foo":1}

to me keeping `1` as `1` would be `the original value` in:

> i thinks it's more error prone than having `the original value`

antonmes:
let's move to slack

czechboy0:
I want to make sure I refute some frequently repeated false assumptions
- "JSON spec defines double" - it doesn't, it says the library is in charge of representation, which I'm proposing (by giving the library enough precision to make the right call)
- "Apple mixes int and double" - they have to, because they use NSNumber. They also mix Booleans in, so are you suggesting we don't parse bool separately from numbers?
- "It's too much work to have int and double separate" - it's not, as you'll never work with it directly, just as you don't work with it directly now, you use the library's convenience initializers and getters, so no, there's no downside
- "JSON is Javascript, which only has double" - JSON **is not** Javascript, it's a notation with a similar syntax. JSON is to Javascript as much as Javascript is to Java.

I still don't understand why some people are against parsing as much type information as possible, when we have a chance to nicely represent it in Swift. There is not `Number` type, so we have to represent is in a way that makes sense in Swift. I'd be fine with @paulofaria's suggestion for adding a Number type (or the way I proposed it, either is fine with me). 

We **need** to make a decision of how we represent a number. I am for int and double, depending on what we parse, because Int's and Double's are **very** different representations, so mixing them would always make the other less precise (for no benefit).

In Swift, we **like** type information, it makes the code safer.

Once again - with my proposal, both our and your way of usage **is possible**. Both parties can be happy. With yours, only you are happy, our side is suddenly robbed of type information, which would be very useful for higher strictness.

To pull it all together - please understand that I wrote a JSON library, read the JSON spec and spent quite some time thinking about it and testing it, and this feels like the best way to represent a number in a world where we can easily tell them apart. I prefer to give my users the maximum information that I parsed, so that they can use it, because Double/Int/Bool/String are all **very** different types to me (and most people). Just because on the web they sometimes get mushed together doesn't mean **all** applications do that.

If you're really so against this proposal, which I don't understand, as there are many more _more_ disputable things in C7 that never had a peer review of this magnitude, I guess I'll have to step back and close this PR. I can't in good faith mix Doubles and Ints and call myself a computer scientist. If someone else wants to do it, go ahead, but I won't be behind a standard like that.

Once again, with my proposition, we **can make everyone happy**. With yours, **only you** will be happy. That alone should be something to think about.

I'd propose this still gets merged, and if we **encounter** real problems (which I don't believe we will), we can change it, I'd be fine with that. Unfortunately all the previous conversation was about very hypothetical cases, that I've never seen any actual evidence happen. If we do merge it, I already have two JSON libs that both can benefit greatly from being compatible, and down the road users will benefit.

Don't be selfish, if we can make both groups happy, I don't see why we can't.

antonmes:
as you said https://github.com/open-swift/C7/pull/32#issuecomment-219579822

> Not the same thing, as if I send 1.0, it's a double but this function would actually claim it's an int. Just because something is convertible to int doesn't mean it was intended to be used as int. But just read the whole thread, I've explained my reasoning.

and with current implementation we would have integer while client was sending float. https://github.com/open-swift/C7/pull/32#issuecomment-219605570

> open devtools in your browser and paste it into the console: `JSON.stringify({num: 1.0})`

so we have to handle both cases everywhere in our code.
you're saying you worry about losing type information, but we just literally can't know was it double or integer on the client side, so the parser makes an assumption that it was an integer. Because `12.00` would be serialized as `12`. And you can do the same check later on e.g. `if value % 1.0 == 0.0`
how is it better making assumptions on parser level vs checking values with `optional properties`?
this is only issue i worried about. not those assumptions you mention: the type is already lost. 
(@paulo, by saying `type` i mean int vs double. not number 😜)

antonmes:
``` swift
extension JSON {
    var integer: Int? {
        if case .number(let value) = self where value % 1.0 == 0.0 {
            return Int(value)
        }
        return nil
    }
}
```

antonmes:
``` swift
public enum JSON {
    case object([String: JSON])
    case array([JSON])
    case number(Double)
    case string(String)
    case boolean(Bool)
    case null
}

extension JSON {
    var integer: Int? {
        if case .number(let value) = self where value % 1.0 == 0.0 {
            return Int(value)
        }
        return nil
    }
}

var answer = JSON.number(42)
var pi = JSON.number(3.14)

if let _ = answer.integer {
    print("the answer")
}

guard let _ = pi.integer else {
    print("pi zdec")
    exit(0)
}
```

czechboy0:
@antonmes I don't think I understand.

You're talking about what some Javascript parsers do - they can only do what their language allows them to. We have a better language that allows us to discern between the two, so I think we should.

If you send me "1.0", I'll parse it as Double, if you send me "1", I'll parse it as Int. Why is this bad?

czechboy0:
@antonmes You also gave me the perfect example of how bad this looks

You're suggesting:

```
public enum JSON {
    case object([String: JSON])
    case array([JSON])
    case number(Double)
    case string(String)
    case boolean(Bool)
    case null
}
```

Let's read it from the top:
- object -> dictionary (cool)
- array -> array (cool)
- number -> double? (what? why?)
- string -> string (cool)
- boolean -> bool (cool)

Once again, since Swift doesn't have a Number type, we need to create one for this purpose. And two fundamental ways to represent numbers are Int and Double.

I think I'm done explaining this. If anyone is unsure, read the whole conversation.

I believe my proposal is in line with the goals of the Swift language much more than mixing doubles and ints, so I'm going to keep this. 
Even if you're not 100% happy with how it looks, the way you want to use it is still possible (mixing doubles and ints, with the help of a library), thus my proposal doesn't prevent you from doing anything (your modifications would prevent me, though). And once again, nobody will use this type directly except for convenience libraries and parsers, so the point is moot.

In light of this, I'll merge this the way it is. If you can gather actual information (data, not opinions) about how this is actively bad, I'll listen. But this cannot be a veto-based process, I just asked you all for feedback. Thanks for it, I hope I answered all your concerns. Even if you don't fully agree, I explained why I think this is the best solution and in order to get anywhere, we need to build things, not spend days on talking about something that can only be properly understood by being used in the wild.

I hope I can convince you about this being the right move in the following days, when I'm planning to make two different libraries compatible with this new standard Swift JSON, and you'll see how beautifully they can work together. This is why I started this and I believe the benefits will hugely outweigh any downsides.

Thanks everyone, it was exhausting, but I'm glad everyone had a chance to give their opinion. If you come up with something new (not already mentioned above, please read the full thread first), just comment below and I'll try to react.

To the future! 💯

paulofaria:
I agree with you on everything except the forced merge. I reverted the commit. We should keep discussing this until we reach a consensus. I'll create another PR with your branch because this one is lost.

czechboy0:
Feel free to keep discussing, but I'm out. Unfortunately the discussion
isn't about the technical points anymore and basic computer science
principles are ignored, so I'll go work on something else instead. Let me
know if this actually ever gets merged, and I'll be happy to reconsider.

PR #39, opened 2016-05-18
RFC 7159 JSON Using Int and Double

bradhilton:
For those of us who want something simple and Swifty:

``` swift
public enum JSON {
    case object([String: JSON])
    case array([JSON])
    case integer(Int)
    case double(Double)
    case string(String)
    case boolean(Bool)
    case null
}
```

Per Apple's official _The Swift Programming Language_ book, `Int` and `Double` are the most fundamental number types for Swift users and the most appropriate choice:

> Unless you need to work with a specific size of integer, always use Int for integer values in your code.
> 
> The appropriate floating-point type to use depends on the nature and range of values you need to work with in your code. In situations where either type would be appropriate, Double is preferred.

### DIFF
--- a/Sources/JSON.swift
+++ b/Sources/JSON.swift
@@ -1,0 +1,9 @@
+public enum JSON {
+    case object([String: JSON])
+    case array([JSON])
+    case integer(Int)
+    case double(Double)
+    case string(String)
+    case boolean(Bool)
+    case null
+}


antonmes:
https://github.com/open-swift/C7/pull/36#issuecomment-220512095
i like this one as well, but only if we add UInt and convenience [initializers](https://github.com/open-swift/C7/pull/38/commits/bad69ed399aa5a6864ca2c8b04f359de1854e514)

antonmes:
but.. this one comes with another trade-off: 
we'll need a property like  `isNumber` and have to write 2 more `cases`,
while we could move/ignore this logic

paulofaria:

PR #42, opened 2016-05-26
fuzzier json number equatable

loganwright:
One potential implementation of fuzzier json number equality. When creating a double, I don't know how to check if below Double max similar to others if someone has suggestions.

### DIFF
--- a/Sources/JSON.swift
+++ b/Sources/JSON.swift
@@ -13,7 +13,6 @@ public enum JSON {
 }
 
 extension JSON.Number {
-    // TODO: We could make these public, but people might prefer not to.
     private var double: Double {
         switch self {
         case let .double(d):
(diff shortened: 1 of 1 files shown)


Sources/JSON.swift
```diff
@@ -12,13 +12,57 @@ public enum JSON {
...
+            } else {
+                return nil
+            }
+        }
+    }
+
+    private var uint: UInt? {
+        switch self {
+        case let .double(d) where d >= 0:
+            return UInt(d)
```
czechboy0, line 47: Doesn't this cut the floating point part of the number?

```
let d = 2.1
UInt(d) //2
```

czechboy0, line 47: So if you'd be comparing `UInt(2)` and `Double(2.1)`, it'd return `true`. That doesn't seem right

loganwright:
@czechboy0 totally right, just updated!

czechboy0:
Cool, just left one comment for something that I think should be fixed. Once that's done I'll merge, thanks @LoganWright! (Even though it's not 100% correct (Int.max gets cut when converted to Double), but 99.99% might be good enough here, it definitely is easier to use)

PR #46, opened 2016-06-14
Swift 3.0 Preview 1

Danappelxx:
Not sure what to set `.swift-version` to. After installing Xcode 8.0, `swiftenv versions` returns `3.0` as the latest version. However, this will probably not work on Ubuntu, where it is `swift-3.0-preview-1`.

Lets get this merged asap regardless :)


czechboy0:
And get all CI to break from now on? That doesn't sound like a great idea 😊

Danappelxx:
Yeah, we're kind of stuck there if we want to use swiftenv. I opened kylef/swiftenv#37 to hopefully address it.

kylef:
I'd suggest setting the `SWIFT_VERSION` environment variable on OS X to `3.0`. Then set the `.swift-version` file to `3.0-preview-1` (because that is the actual version you're supporting at the moment).

In CI, you could perhaps do the following:

``` yaml
install:
- if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export SWIFT_VERSION="3.0"; fi
- eval "$(curl -sL https://gist.githubusercontent.com/kylef/5c0475ff02b7c7671d2a/raw/9f442512a46d7a2af7b850d65a7e9bd31edfb09b/swiftenv-install.sh)"
```

kylef:
Actually, since Swift on OS X is included with Xcode. You will need to install the version of Xcode that bundles this version of Swift on Travis CI. There is no binary snapshots so you can't `swiftenv install` it.

Danappelxx:
Thanks @kylef! I think what I updated it to works (CI passes).

